PR #66, opened 2026-03-21
Fix workspace QA session mode badges

bbsngg:
## Summary
- persist session mode metadata when creating new Cursor, Codex, Gemini, and Claude sessions
- restore Cursor and Codex session modes from indexed metadata when building project session lists
- improve Claude and Codex fallback parsing so older sessions recover mode from explicit session-mode context in the first real user message
- improve Gemini fallback recovery so older sessions can recover workspace QA mode from original tmp chat context when indexed metadata is missing or wrong
- prefer stored workspace QA mode in the sidebar badge fallback so historical sessions keep the correct label

## Verification
- node --check server/projects.js
- node --check server/cursor-cli.js
- node --check server/openai-codex.js
- node --check server/gemini-cli.js
- node --check server/claude-sdk.js
- locally verified Codex session 019d0e8c-dda8-7be2-bbc3-533da4919a21 now resolves to mode workspace_qa
- locally verified historical Gemini session a257b1f1-2620-4a37-bb60-36af39d8f12b now resolves to mode workspace_qa
- locally verified historical Claude session 9651ee21-4a62-4122-b73a-0681b5e38dba now resolves to mode workspace_qa
- restarted the app and confirmed the workspace QA badge renders correctly

### DIFF
--- a/server/claude-sdk.js
+++ b/server/claude-sdk.js
@@ -18,18 +18,59 @@ import { promises as fs } from 'fs';
 import path from 'path';
 import os from 'os';
 import { CLAUDE_MODELS } from '../shared/modelConstants.js';
-import { ensureProjectSkillLinks } from './projects.js';
+import { encodeProjectPath, ensureProjectSkillLinks, reconcileClaudeSessionIndex } from './projects.js';
 import { writeProjectTemplates } from './templates/index.js';
 import { recordIndexedSession } from './utils/sessionIndex.js';
 
 import { createRequestId, waitForToolApproval, resolveToolApproval as resolvePermApproval, matchesToolPermission } from './utils/permissions.js';
 
 const activeSessions = new Map();
+const pendingClaudeSessionIndexReconciles = new Map();
 
 const TOOLS_REQUIRING_INTERACTION = new Set(['AskUserQuestion']);
 
 function resolveToolApproval(requestId, decision) {
   resolvePermApproval(requestId, decision);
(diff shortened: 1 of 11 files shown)


bbsngg:
Addressed in `d696b14` (on top of the `main` merge in `10884fe`).

Fixed in this pass:
- Restored `getSessionMode` precedence so the UI now trusts `session.mode` first and only falls back to localStorage when the server value is missing.
- Tightened `inferSessionModeFromUserMessage` to only use high-confidence workspace signals (`git` / `commit` / `push` / `pull request` / `repo` etc.), which avoids false positives on generic coding prompts like `function`, `test`, or `file`.
- Removed the old Gemini tmp recursive lookup from the active recovery path and documented the effective recovery order in code: DB metadata -> JSONL metadata -> explicit context marker -> high-confidence heuristic -> default.
- Normalized explicit metadata parsing so `readExplicitSessionModeFromMetadata` now handles casing consistently.
- Deduplicated the bootstrap Gemini metadata entry so we only persist `payload.sessionMode` instead of writing the same mode value in multiple places.

Also covered with verification:
- `node --test test/session-mode.test.mjs`
- `npm run typecheck`
- `git diff --check`

Separately, after merging `main`, the provider-side session indexing path is already unified through `recordIndexedSession()`. The path-collision concern in `encodeProjectPath` is still a separate follow-up if we want to harden the encoding scheme further.

PR #69, opened 2026-03-21
Fix Xiaohongshu login flow for remote deployments

bbsngg:
## Summary
- clarify that browser cookie extraction runs on the machine hosting the dr-claw service
- add browser selection and QR-code login as Xiaohongshu auth options
- default to highlighting QR login for likely remote-access deployments and surface clearer login feedback

## Testing
- npm run typecheck
- npm run build


Zhang-Henry:
## Code Review

### HIGH

1. **Server/frontend browser list mismatch** — Server `allowedCookieSources` includes `safari` and `opera_gx`, but the frontend `XHS_COOKIE_SOURCES` dropdown omits both. Users can never select these options even though the server accepts them. Either add them to the frontend or remove from the server allowlist.

### MEDIUM

2. **Double-response crash risk** — Both `child.on('close')` and `child.on('error')` can fire on the spawned process, each calling `res.json()` / `res.status(500).json()`. The second call throws `ERR_HTTP_HEADERS_SENT`. Add a `let responded = false` guard in each handler.

3. **Korean locale has many untranslated keys** — `cookieSource`, `extractBrowserCookie`, `loginViaQr`, `qrLoggingIn`, `qrSuccess`, `qrFailure`, `recommendedTag` are all still in English, while zh-CN is fully translated.

4. **`contextHint` is a static string** — It's determined before the child process runs and returned verbatim regardless of success or failure. Consider only sending the hint on failure, or making it context-aware.

### LOW

5. Browser Source dropdown is always visible even when QR login method is selected — consider conditionally rendering it only for browser-cookie login.
6. `handleLogin` sends `cookieSource` in the request body even for `method: 'qrcode'` — server ignores it, but adds noise for debugging.
7. QR login button uses a `Terminal` icon — consider using `QrCode` from lucide-react for better semantic clarity.

PR #72, opened 2026-03-22
Fix session index reconciliation and stuck loading state

bbsngg:
## Summary
- split session index placeholder writes from source-of-truth reconciliation so Claude, Gemini, and Codex sessions can refresh names, counts, timestamps, and real IDs safely
- move Claude completion cleanup off the critical path and reconcile single-session files instead of rescanning an entire project before sending `claude-complete`
- stop the chat UI from treating stale `Resuming...` timer state as an active run, while keeping a short resume/status-check grace window

## Testing
- `node --check server/projects.js`
- `node --check server/claude-sdk.js`
- `npm run typecheck`
- validated stale Claude session metadata is reconciled from real `.jsonl` data in the local DB
- validated affected session records now show synced names/counts/timestamps in `session_metadata`

### DIFF
--- a/server/gemini-cli.js
+++ b/server/gemini-cli.js
@@ -1127,6 +1127,9 @@ export async function spawnGemini(command, options = {}, ws) {
       if (sessionData?.heartbeat) clearInterval(sessionData.heartbeat);
       activeGeminiSessions.delete(finalSessionId);
       await cleanupGeminiTempFiles(tempFilePaths, tempDir);
+      // Send completion event immediately so the UI can settle
+      ws.send({ type: 'gemini-complete', sessionId: finalSessionId, exitCode: code, isNewSession: (!sessionId || sessionId.startsWith('new-session-')) && !!command });
+      // Post-completion housekeeping — runs after the UI receives the completion signal
       if (workingDir && finalSessionId) {
         try {
           await reconcileGeminiSessionIndex(workingDir, {
@@ -1137,7 +1140,6 @@ export async function spawnGemini(command, options = {}, ws) {
           console.warn(`[Gemini] Failed to reconcile indexed session ${finalSessionId}:`, error.message);
         }
       }
-      ws.send({ type: 'gemini-complete', sessionId: finalSessionId, exitCode: code, isNewSession: (!sessionId || sessionId.startsWith('new-session-')) && !!command });
       if (policyViolationTriggered || code === 0 || code === null) resolve();
       else reject(new Error(`Gemini CLI exited with code ${code}`));
     });
--- a/server/openai-codex.js
+++ b/server/openai-codex.js
@@ -391,6 +391,15 @@ export async function queryCodex(command, options = {}, ws) {
     }
 
     const actualSessionId = thread.id || currentSessionId;
+
+    // Send completion event immediately so the UI can settle
+    sendMessage(ws, {
+      type: 'codex-complete',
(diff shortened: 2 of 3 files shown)


bbsngg (commented):
## PR Review Summary

Overall this is a solid improvement that addresses real UX issues (stuck loading) and data consistency problems (placeholder vs source-of-truth separation). The architecture direction is sound. A few items worth addressing:

### Issues

**1. Dead code in `reconcileClaudeSessionIndex` (server/projects.js)**
When `targetSessionId` is null, the function calls `getSessions(projectName, 0, 0)` then immediately hits `if (!targetSessionId) return result;` — the code block below it (lines ~708-711 with the `.session` field) is unreachable.

**2. Duplicated timestamp normalization logic**
- `toIsoString()` in `projects.js` duplicates `normalizeSessionTimestamp()` in `db.js`
- `normalizeSessionTimestamp` and `normalizeSessionCreatedAt` in `db.js` are nearly identical (only difference is the `replace('T', ' ').slice(0, 19)` formatting)

Consider consolidating to avoid behavioral drift.

**3. Hardcoded magic values**
- The 5000ms timeout in `useChatSessionState.ts` for clearing stale loading state should be extracted to a named constant with a comment explaining the chosen duration. If WebSocket reconnection takes longer than 5s, this could prematurely clear a legitimately active session's loading state.
- The string `'Resuming...'` appears in both `ChatInterface.tsx` and `useChatSessionState.ts` — should be a shared constant.

**4. Sequential awaits in Gemini/Codex reconcile loops (server/projects.js)**
The `for...of` + `await` pattern in `getGeminiSessions` and `getCodexSessions` sync paths will serialize DB writes. Consider `Promise.all`/`Promise.allSettled` for parallelism when session count is non-trivial.

### Minor / Non-blocking

- Sidebar sort order change (`lastActivity || createdAt` instead of `createdAt || lastActivity`) for codex/gemini is a behavior change — makes sense (sort by recent activity), just confirming it's intentional.
- `setTimeout(async () => ...)` in `scheduleClaudeSessionIndexReconcile` — the try/catch covers it, but unhandled rejections inside setTimeout callbacks can be subtle; current code is fine.
- Error-path reconcile in `claude-sdk.js` catch block uses `Promise.allSettled` which silently swallows reconcile failures — acceptable since it's best-effort cleanup.

### Verdict

Good to merge after addressing items 1-3. Item 4 is a nice-to-have optimization.

bbsngg:
Pushed `15689d0` to address the review feedback:

1. **Dead code removed** — cleaned up unreachable `if (!targetSessionId)` guard and dead return in `reconcileClaudeSessionIndex`
2. **Timestamp normalization consolidated** — removed `toIsoString()` from `projects.js`, now reuses `normalizeSessionTimestamp` exported from `db.js`; added comment to `normalizeSessionCreatedAt` explaining the `YYYY-MM-DD HH:MM:SS` format for SQLite convention
3. **Magic values extracted** — `RESUMING_STATUS_TEXT` shared constant in `types.ts` (replaced 5 occurrences across 3 files), `STATUS_VALIDATION_TIMEOUT_MS` constant with doc comment in `useChatSessionState.ts`
4. **Reconcile loops parallelized** — Gemini and Codex `syncIndex` blocks now use `Promise.allSettled` instead of sequential `for...of`

All checks pass: `node --check server/projects.js`, `node --check server/database/db.js`, `npm run typecheck`.

PR #79, opened 2026-03-23
Fix Codex discovery without Claude metadata

bbsngg:
## Summary
- detect Codex authentication from the server environment `OPENAI_API_KEY` as well as `~/.codex/auth.json`
- discover and index Codex-only projects directly from `~/.codex/sessions` so they appear without prior Claude metadata
- add regression tests for Codex auth fallback and Codex-only project candidate extraction

## Testing
- `$(dirname "$(command -v codex)")/node --test test/codex-discovery.test.mjs`
- manual integration check: temporary home with only `~/.codex/sessions` and no `.claude`, then `getProjects()` returns the project with `codexSessions`


Zhang-Henry (commented):
## Code Review

Overall: Well-structured PR solving a real gap — Codex-only projects (no `.claude` metadata) couldn't be discovered. Clean code, tests included. Main concerns are performance and a missing export.

### Issues

**1. [Performance] `syncDiscoveredProjectsFromCodexSessions` runs on every `getProjects()` call**

Previously, `bootstrapProjectsIndexFromLegacySources` only ran when `dbProjects.length === 0` (first launch). Now `syncDiscoveredProjectsFromCodexSessions` executes **unconditionally**, and internally it calls `buildCodexSessionsIndex()` which:

1. `findCodexJsonlFiles(codexSessionsDir)` — recursive directory traversal of `~/.codex/sessions/`
2. `parseCodexSessionFile()` for each `.jsonl` file — readline line-by-line parsing
3. `normalizeComparablePath()` for each path

`getProjects()` is called on every sidebar render. For users with many Codex session files, this creates significant I/O overhead on every render cycle.

Suggested fixes (pick one):
- Add caching with TTL (e.g., 30-second cooldown between rescans)
- Use a `last_synced` timestamp check
- Simplest: move inside the `if (dbProjects.length === 0)` block to only sync on first launch

**2. [Bug] `encodeProjectPath` is not exported but imported in tests**

The test file imports:
```js
import { collectCodexProjectCandidates, encodeProjectPath } from '../server/projects.js';
```

But checking the main branch export list (`projects.js:3559-3587`), `encodeProjectPath` is **not exported**. The PR diff only adds `collectCodexProjectCandidates` to exports. The test will fail at import time.

Fix: add `encodeProjectPath` to the export block.

**3. [Minor] Auth precedence change worth documenting**

`checkCodexCredentials` now checks in this order:
1. JWT tokens in `auth.json`
2. **Environment variable `OPENAI_API_KEY`** (new)
3. `OPENAI_API_KEY` in `auth.json` file

The env var taking priority before `auth.json`'s API key is reasonable for service deployments but changes existing behavior. Worth a brief code comment documenting the precedence order.

**4. [Minor] Test uses `process.execPath` as Codex CLI stand-in**

```js
process.env.CODEX_CLI_PATH = process.execPath; // Node binary
```

The Node binary passes `resolveAvailableCliCommand`'s existence check but isn't actually Codex. If `resolveAvailableCliCommand` adds version verification in the future, this test will break. Consider adding a comment explaining why this works.

### Strengths

- `collectCodexProjectCandidates` is a pure function cleanly separated from side effects — easy to test and reason about
- Session deduplication by ID within a candidate is correct
- `withTempHome` test helper is well-written with proper environment save/restore and `t.after()` cleanup
- `auth.json` read changed to graceful ENOENT handling — good backwards compatibility
- `upsertSessionFromSource` in db.js uses smart merge strategy (latest timestamp, metadata merge, message count resolution)

PR #82, opened 2026-03-23
feat(chat): show Running code status during Bash execution

MarkSiqiZhang:
## Summary                                                                                                                                          
   - When an agent executes Bash/shell commands, the status now shows **"Running code"** instead of generic "Thinking"                               
   - Applies to all 3 providers: Claude, Codex, and Gemini
   - Other tools (Edit, Grep, WebSearch, etc.) continue to show "Thinking"

## Changes
   - **`useChatRealtimeHandlers.ts`** — Detect `Bash`/`run_shell_command` tool_use and update `claudeStatus.text` to "Running code" for
   Claude/Gemini; detect `command_execution` for Codex
   - **`AssistantThinkingIndicator.tsx`** — Accept `statusText` prop, default to i18n `status.thinking`
   - **`ChatMessagesPane.tsx`** — Thread `statusText` prop to indicator
   - **`ChatInterface.tsx`** — Pass `claudeStatus?.text` to ChatMessagesPane
   - **i18n** — All 3 locales (en, zh-CN, ko) updated with `status.thinking` and `status.runningCode`

## Test plan
   - [ ] Send a message that triggers Bash execution → status bar and thinking indicator show "Running code"
   - [ ] Send a normal message (no tool use) → status shows as before
   - [ ] Edit/Write/Grep tools → status does not stay "Running code"
   - [ ] Works for Claude, Codex, and Gemini

### DIFF
--- a/src/components/chat/hooks/useChatRealtimeHandlers.ts
+++ b/src/components/chat/hooks/useChatRealtimeHandlers.ts
@@ -14,6 +14,7 @@ import {
   safeLocalStorage,
 } from '../utils/chatStorage';
 import { RESUMING_STATUS_TEXT } from '../types/types';
+import i18n from '../../../i18n/config';
 import type { ChatMessage, PendingPermissionRequest } from '../types/types';
 import type { Project, ProjectSession, SessionProvider } from '../../../types/app';
 
@@ -49,6 +50,7 @@ interface UseChatRealtimeHandlersArgs {
   setIsLoading: (loading: boolean) => void;
   setCanAbortSession: (canAbort: boolean) => void;
   setClaudeStatus: Dispatch<SetStateAction<{ text: string; tokens: number; can_interrupt: boolean; startTime?: number } | null>>;
+  setStatusTextOverride: Dispatch<SetStateAction<string | null>>;
   setTokenBudget: (budget: Record<string, unknown> | null) => void;
   setIsSystemSessionChange: (isSystemSessionChange: boolean) => void;
   setPendingPermissionRequests: Dispatch<SetStateAction<PendingPermissionRequest[]>>;
@@ -139,6 +141,7 @@ export function useChatRealtimeHandlers({
   setIsLoading,
   setCanAbortSession,
   setClaudeStatus,
+  setStatusTextOverride,
   setTokenBudget,
   setIsSystemSessionChange,
   setPendingPermissionRequests,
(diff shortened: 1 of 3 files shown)


Zhang-Henry (commented):
## Code Review

Overall: Clean, small PR with clear intent. Two functional bugs should be addressed before merge.

### Issues

**1. [Bug] Hardcoded English string bypasses i18n**

`useChatRealtimeHandlers.ts` lines 179-181 and 901 directly set `text: 'Running code'` as a literal English string, but the i18n keys `status.runningCode` are already defined in all 3 locales (`chat.json`). This means:
- Chinese users see "思考中..." → "Running code" (mixed languages)
- Korean users likewise see English

The issue is that `useChatRealtimeHandlers` is a hook where calling `t()` directly isn't straightforward. Suggested fix: instead of setting display text in the hook, set a semantic key (e.g., `text: 'running_code'`), then resolve it via i18n in `AssistantThinkingIndicator` at render time. Currently `status.runningCode` is defined in all 3 locale files but **never referenced** — it's dead code.

**2. [Bug] Status text never resets after Bash execution completes**

Tracing the full lifecycle:
- `claudeStatus` is defined at `useChatSessionState.ts:94` as `{ text, tokens, can_interrupt } | null`
- Set to non-null: when `claude-status`/`gemini-status` messages arrive (`useChatRealtimeHandlers.ts:1134-1147`) — `text` comes from backend's `statusData.message`
- Cleared to null: `clearLoadingIndicators()` on session complete/error/abort (`useChatRealtimeHandlers.ts:392-396`)

The PR sets `text` to `'Running code'` on Bash tool_use detection, but **nothing resets it back**. If Bash finishes and the model continues thinking (e.g., processing results before emitting text), `text` remains `'Running code'` until either the backend pushes a new `claude-status` event or the session ends.

In practice this may be masked by frequent backend status pushes, but in edge cases (Bash completes → model directly outputs text without a new status push), the indicator stays stuck on "Running code".

Suggestion: reset `text` when a tool_result arrives or when a new text/thinking content block begins.

**3. [Minor] Codex `command_execution` triggers on all lifecycle stages**

The `command_execution` case (~line 793) handles all lifecycle phases (started/completed/etc.). The PR sets the status at the case entry, meaning even the `completed` phase triggers "Running code". Should only set on command start.

**4. [Nit] Gemini coverage**

PR description says "Applies to all 3 providers", but Bash detection is only added to Claude and Codex code paths. Gemini shares the `claude-status`/`gemini-status` handler but its tool_use events go through the same path as Claude (shared case), so Bash detection at line 179 should cover Gemini too — worth verifying this is actually the case.

PR #124, opened 2026-04-04
Support OpenRouter-compatible relay base URLs across UI, API, and CLI

xieli123-hub:
## Summary
This PR adds first-class support for OpenRouter-compatible relay / proxy endpoints so Dr. Claw can work with relay-based APIs without assuming the official OpenRouter base URL everywhere.

## What Changed
- centralize OpenRouter base URL normalization and official-host header handling
- let Settings > OpenRouter verify and persist both the API key and a custom relay/base URL
- pass the configured base URL through UI chat sessions, the external /api/agent flow, and dr-claw chat --base-url
- fetch and cache model lists by base URL instead of assuming the official endpoint
- add regression coverage for base URL parsing and CLI argument handling
- document relay setup in .env, CLI usage, and the external API docs

## Verification
- 
pm.cmd run typecheck
- 
pm.cmd run test -- server/__tests__/openrouter-config.test.mjs server/__tests__/cli.test.mjs

## ????
- ?? PR ? OpenRouter ??????? / ????,???????????? OpenRouter ???
- UI ??????????? /api/agent ??? CLI ?????? base URL?
- ???????????? base URL ??,???? OpenRouter ? OpenRouter-compatible ?????
- ?????????????,??????? OpenRouter ??????? API?

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -47,6 +47,7 @@ VITE_CONTEXT_WINDOW=160000
 # =============================================================================
 # OPENAI_API_KEY=sk-...
 # OPENROUTER_API_KEY=sk-or-...
+# OPENROUTER_BASE_URL=https://openrouter.ai/api/v1
 # OPENROUTER_MODEL=anthropic/claude-sonnet-4
 
 # =============================================================================
--- a/README.md
+++ b/README.md
@@ -521,6 +521,9 @@ Auto Research email notifications are configured inside the app at **Settings â†
    - **Environment variable:** `export OPENROUTER_API_KEY=sk-or-...`
    - **`.env` file:** add `OPENROUTER_API_KEY=sk-or-...` to your project `.env`
    - **UI:** go to **Settings â†’ OpenRouter** and paste your key
+3. If you use a relay / proxy endpoint, also set `OPENROUTER_BASE_URL` to the full compatible base path:
+   - **Official OpenRouter:** `OPENROUTER_BASE_URL=https://openrouter.ai/api/v1`
+   - **Relay example:** `OPENROUTER_BASE_URL=https://your-relay.example.com/v1`
 
 ### Using OpenRouter in the UI
 
@@ -541,6 +544,9 @@ node server/cli.js chat --model moonshotai/kimi-k2.5
 
 # With an explicit API key
 node server/cli.js chat --model deepseek/deepseek-r1 --key sk-or-your-key
+
+# With a relay / proxy endpoint
+node server/cli.js chat --model deepseek/deepseek-r1 --base-url https://your-relay.example.com/v1
 ```
 
 The CLI supports the same tools as the UI (file I/O, shell, grep, glob, web search, web fetch, todo). Type your message and the agent will execute multi-step research tasks autonomously.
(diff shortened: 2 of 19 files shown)


Zhang-Henry (commented):
## Code Review

Clean central architecture — `openrouterConfig.js` normalizes base URLs and conditions provider-specific headers across all three callsites. The utility design is solid and well-tested. However, there are real issues to address before merging.

---

### Blocking

**1. SSRF via user-controlled `baseUrl` in the agent API route**
`server/routes/agent.js`: The `baseUrl` from `req.body` is passed directly to `queryOpenRouter`, which uses it as the target of a `fetch`. Unlike `/openrouter/verify-api-key` (which validates via `normalizeOpenRouterBaseUrl`), the agent route performs no validation. A caller with a valid API key can direct the server to fetch any internal URL (`http://169.254.169.254/`, `http://localhost:3000/admin`, etc.) and have the response streamed back via WebSocket.

Fix: validate that the hostname is not a private/loopback/link-local address, or restrict `baseUrl` to a server-side env setting only.

---

### Should Fix

**2. Default base URL always written to `.env` on key save**
`server/routes/cli-auth.js` (~L603): `normalizeOpenRouterBaseUrl` returns the official URL when `req.body.baseUrl` is undefined/empty. The code then unconditionally writes `OPENROUTER_BASE_URL=https://openrouter.ai/api/v1` to `.env` and sets `process.env.OPENROUTER_BASE_URL`. Users who save an API key without filling in a relay URL get a surprise `.env` entry. Fix: only write/set `OPENROUTER_BASE_URL` when the caller explicitly supplied a non-empty value.

**3. `agentLoop` uses `baseUrl` via closure, not parameter**
`server/cli-chat.js` (~L412): `agentLoop` references `baseUrl` as a closure variable from `startChat`. If the function is ever called from elsewhere or moved, `baseUrl` will be undefined. The signature should explicitly receive `baseUrl` as a parameter.

**4. `.env` parser is fragile**
`server/routes/cli-auth.js`, `upsertEnvValues`: `.split('=')[0]` works for simple cases but doesn't handle edge cases like whitespace-prefixed keys (` OPENROUTER_API_KEY=...`). Consider using a proper `.env` line-parse regex or a library like `dotenv-edit`.

---

### Minor

- Raw fetch error messages are echoed to the API caller in the catch block (~L631) — could leak internal URL info if a relay returns verbose errors. Consider sanitizing.
- The `filter` logic for trailing blank lines in `upsertEnvValues` doesn't de-duplicate consecutive blank lines in the middle of the file.
- Removing the module-level `_modelsCache` in `ProviderSelectionEmptyState.tsx` means every remount triggers an API call to `/api/settings/openrouter-models`. The server-side 30-minute cache absorbs this, but worth noting.

### Positive Notes

- The `openrouterConfig.js` separation of concerns (`normalizeOpenRouterBaseUrl` throws on bad input, `getOpenRouterBaseUrl` is safe/fallback, `isOfficialOpenRouterBaseUrl`, `getOpenRouterProviderHeaders`) is clean and testable.
- `isDirectExecution` guard in `server/cli.js` is the right approach for ESM test imports.
- `closeDatabase` addition to `db.js` and its use in tests is proper cleanup.
- Model-list API fallback for flat JSON arrays vs `{ data: [...] }` is good defensive coding.
- Test coverage for `openrouterConfig.js` and `cliArgs.js` is thorough.

Zhang-Henry:
@xieli123-hub

PR #135, opened 2026-04-06
fix(security): harden shell execution, CORS, and path validation

haoyu-haoyu:
## Summary

Four security fixes addressing shell injection, path traversal, and CORS misconfiguration across 5 files.

### 1. Shell injection in `server/routes/user.js` (Critical)

**Before:** Git user name/email interpolated into shell strings via `execAsync`:
```js
await execAsync(`git config --global user.name "${gitName.replace(/"/g, '\\"')}"`);
```
A name like `` `whoami` `` or `$(rm -rf /)` bypasses the double-quote escaping.

**After:** Uses `spawnAsync` with `shell: false` and argument arrays — no shell parsing, no injection vector. Same pattern as our merged PR #88 for `server/routes/git.js`.

### 2. Shell injection in `server/openrouter.js` + `server/cli-chat.js` (Critical)

**Before:** LLM-driven tool calls (Glob, Grep) passed `args.pattern` directly into shell commands:
```js
`rg --files --glob '${args.pattern}'`
```

**After:**
- All tool handlers use `spawnAsync` with argument arrays (`shell: false`)
- Grep uses `-e pattern -- target` to prevent argv option injection
- `spawnAsync` enforces 1MB `maxBuffer` with SIGTERM to prevent DoS from large tree output

### 3. Path traversal in `server/routes/news.js` (Medium)

**Before:** `/api/news/logs/:source` used `req.params.source` in `path.join` without validation.

**After:** Added `getSourceEntry()` check before path construction, matching the validation pattern used by all other `:source` endpoints in the same file. Returns 404 for unknown sources.

### 4. CORS misconfiguration in `server/index.js` (Medium)

**Before:** `cors()` with no restrictions — any origin can make credentialed requests to `localhost:3001`.

**After:** CORS restricted to `DR_CLAW_CORS_ORIGINS` env var (comma-separated), defaulting to `localhost` and `127.0.0.1` on the frontend port. Empty entries are filtered.

## Testing

- Verified `user.js` git config still works with normal names
- Verified Glob/Grep handlers produce correct output format
- Verified news endpoint returns 404 for invalid sources
- Codex code review completed — addressed argv injection and maxBuffer regression

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -421,7 +421,7 @@ app.locals.wss = wss;
 
 app.use(cors({
   origin: process.env.DR_CLAW_CORS_ORIGINS
-    ? process.env.DR_CLAW_CORS_ORIGINS.split(',').map((o) => o.trim())
+    ? process.env.DR_CLAW_CORS_ORIGINS.split(',').map((o) => o.trim()).filter(Boolean)
     : [`http://localhost:${getFrontendPortSync()}`, `http://127.0.0.1:${getFrontendPortSync()}`],
   credentials: true,
 }));
--- a/server/index.js
+++ b/server/index.js
@@ -419,10 +419,12 @@ const wss = new WebSocketServer({
 // Make WebSocket server available to routes
 app.locals.wss = wss;
 
+// Parse VITE_PORT early so CORS reflects the actual frontend port.
+const CORS_VITE_PORT = parsePortNumber(process.env.VITE_PORT, DEFAULT_FRONTEND_PORT);
 app.use(cors({
   origin: process.env.DR_CLAW_CORS_ORIGINS
     ? process.env.DR_CLAW_CORS_ORIGINS.split(',').map((o) => o.trim()).filter(Boolean)
-    : [`http://localhost:${getFrontendPortSync()}`, `http://127.0.0.1:${getFrontendPortSync()}`],
+    : [`http://localhost:${getFrontendPortSync(CORS_VITE_PORT)}`, `http://127.0.0.1:${getFrontendPortSync(CORS_VITE_PORT)}`],
   credentials: true,
 }));
 app.use(express.json({
--- a/server/index.js
+++ b/server/index.js
@@ -419,7 +419,12 @@ const wss = new WebSocketServer({
 // Make WebSocket server available to routes
 app.locals.wss = wss;
 
-app.use(cors());
+app.use(cors({
+  origin: process.env.DR_CLAW_CORS_ORIGINS
+    ? process.env.DR_CLAW_CORS_ORIGINS.split(',').map((o) => o.trim())
+    : [`http://localhost:${getFrontendPortSync()}`, `http://127.0.0.1:${getFrontendPortSync()}`],
+  credentials: true,
+}));
 app.use(express.json({
   limit: '50mb',
   type: (req) => {
--- a/server/openrouter.js
+++ b/server/openrouter.js
@@ -24,6 +24,7 @@ import { createRequestId, waitForToolApproval, matchesToolPermission } from './u
 const execAsync = promisify(exec);
 
 function spawnAsync(command, args, options = {}) {
+  const maxBuffer = options.maxBuffer || 1024 * 1024; // 1 MB default
   return new Promise((resolve, reject) => {
     const child = spawn(command, args, {
       ...options,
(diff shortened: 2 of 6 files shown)


server/openrouter.js
```diff
@@ -276,25 +327,33 @@ async function executeTool(name, args, workingDir) {
...
-            cwd: dir, timeout: 30_000, maxBuffer: 1024 * 1024,
-          }),
-        );
-        return trunc(stdout || '(no matches)');
+        try {
+          const { stdout } = await spawnAsync('rg', ['--files', '--glob', args.pattern], {
+            cwd: dir, timeout: 30_000,
+          });
+          const lines = stdout.split('\n').filter(Boolean).slice(0, 300);
+          return trunc(lines.join('\n') || '(no matches)');
```
Zhang-Henry, line 285: **Bug: Glob silently returns `"(no matches)"` on large repos — regression from removing `| head -300`**

The old code had `| head -300` in the shell pipeline, which limited `rg` output to ~300 lines via SIGPIPE *before* it hit the buffer. The new code removes that limit — `rg --files` dumps ALL filenames into memory, and only after full collection does JS `.slice(0, 300)`.

For a repo with ~50k files (~2-3MB of path output), this exceeds the default 1MB `maxBuffer` in `spawnAsync`, which triggers:
1. `child.kill('SIGTERM')` → non-zero exit → Promise **rejects**
2. Outer `catch` → falls through to `find` fallback → same overflow → rejects
3. Inner `catch` → `return '(no matches)'`

**Result: Glob appears to work but silently returns empty on any moderately large codebase.**

Same issue in `cli-chat.js` Glob handler.

**Suggested fix:** Make `spawnAsync` **resolve** (not reject) with truncated stdout when `maxBuffer` is exceeded, so the caller still gets the first 1MB of data. Alternatively, limit output at the process level (e.g., pipe through a line counter, or increase maxBuffer for Glob specifically).

server/index.js
```diff
@@ -419,7 +419,12 @@ const wss = new WebSocketServer({
 // Make WebSocket server available to routes
 app.locals.wss = wss;
 
-app.use(cors());
+app.use(cors({
+  origin: process.env.DR_CLAW_CORS_ORIGINS
+    ? process.env.DR_CLAW_CORS_ORIGINS.split(',').map((o) => o.trim()).filter(Boolean)
+    : [`http://localhost:${getFrontendPortSync()}`, `http://127.0.0.1:${getFrontendPortSync()}`],
```
Zhang-Henry: **Bug: `getFrontendPortSync()` called without argument — CORS may block legitimate requests when `VITE_PORT` is customized**

Every other call site in this file passes `REQUESTED_VITE_PORT` as the fallback:
```js
getFrontendPortSync(REQUESTED_VITE_PORT)  // L2898, L3129, L3139
```

Here it's called bare, so the fallback is the hardcoded `DEFAULT_FRONTEND_PORT`. If a user sets `VITE_PORT=3002`, the frontend runs on 3002 but CORS only allows the default port (e.g. 5173). Since the origin array is computed once at middleware setup, it never updates dynamically.

**Note:** `REQUESTED_VITE_PORT` is defined at L3105, *after* this middleware setup runs — so you'll also need to hoist the port variable or move the CORS config later in the file.

server/openrouter.js
```diff
@@ -23,6 +23,57 @@ import { createRequestId, waitForToolApproval, matchesToolPermission } from './u
 
 const execAsync = promisify(exec);
 
+function spawnAsync(command, args, options = {}) {
```
Zhang-Henry: **Code quality: `spawnAsync` is copy-pasted 3 times** (here, `cli-chat.js`, and `user.js`)

The `user.js` copy already diverges — it lacks `maxBuffer` / truncation support. These will drift further apart over time.

Please extract to a shared module (e.g. `server/utils/spawnAsync.js`) and import it from all three files.

Zhang-Henry (commented):
Thanks for the security fixes — the shell injection and CORS issues are real and the `spawnAsync` approach is correct. Found a few bugs that should be addressed before merging.

haoyu-haoyu:
Thanks @Zhang-Henry — all three issues addressed in fca7ba5:

**1. Glob maxBuffer regression**
- Extracted `spawnAsync` to `server/utils/spawnAsync.js`
- On buffer overflow, we `child.kill('SIGTERM')` and the `close` handler checks `truncated && signal === 'SIGTERM'` before resolving — genuine non-zero exits still reject normally
- After truncation, `stdout` is trimmed at the last `\n` so callers never get a phantom partial filename

**2. CORS port bug**
- Added `CORS_VITE_PORT = parsePortNumber(process.env.VITE_PORT, DEFAULT_FRONTEND_PORT)` before the CORS middleware setup
- Now passes `getFrontendPortSync(CORS_VITE_PORT)` so custom `VITE_PORT` values are respected

**3. spawnAsync deduplication**
- Extracted to `server/utils/spawnAsync.js`, imported by all 3 files
- `user.js` now gets maxBuffer/truncation support it was previously missing
- Net -143/+86 lines

PR #136, opened 2026-04-07
fix(security): prevent tool call path traversal outside project root

haoyu-haoyu:
## Summary

Add `safePath()` utility to constrain all LLM-driven tool call paths within the project root, preventing prompt-injection attacks from reading or writing files outside the workspace.

## The vulnerability

When an LLM is prompt-injected (e.g., via a malicious README), it can generate tool calls like `Read({ path: "/etc/passwd" })` or `Read({ path: "../../../.ssh/id_rsa" })`. Since read-only tools (Read, Glob, Grep, LS) are auto-approved without user consent, the file contents are silently exfiltrated through the conversation.

### New: `server/utils/safePath.js`

```javascript
safePath(userPath, allowedRoot) → string | throws Error
```

- **Absolute paths**: Allowed only if they resolve inside the project root
- **`..` traversal**: Resolved and verified to stay within root
- **Symlink escapes**: Resolved via `realpathSync`; for non-existent targets, walks up to the nearest existing ancestor to catch `repo/link/missing/deep/file.txt` where `link -> /external/`
- **Clear error messages**: `"Path traversal blocked: ... resolves to ... which is outside the project root"`

### Modified: `server/openrouter.js` + `server/cli-chat.js`

Replace the unconstrained `resolve()` function in both `executeTool()` implementations:

```javascript
// Before (vulnerable)
const resolve = (p) => path.isAbsolute(p) ? p : path.resolve(workingDir, p);

// After (safe)
const resolve = (p) => safePath(p, workingDir);
```

### Tests: 9 cases covering

- Relative paths within root ✅
- Empty/null/undefined input ✅
- Absolute paths inside root (allowed) ✅
- Absolute paths outside root (blocked) ✅
- `..` traversal above root (blocked) ✅
- Disguised `..` traversal (blocked) ✅
- `..` that stays within root (allowed) ✅
- Non-existent files and nested paths ✅

## Codex review (3 rounds)

1. Initial review → found symlink parent escape and absolute path rejection too strict
2. After fix → found nested missing segments under symlink still escapable
3. After fix → ancestor walk confirmed correct

## Note

This does NOT restrict `Bash` tool execution (which has its own permission model via `permissionMode`). It only constrains file-system tool calls (Read, Write, Edit, Glob, Grep, LS).

### DIFF
--- a/server/utils/safePath.js
+++ b/server/utils/safePath.js
@@ -14,9 +14,12 @@ import fs from 'fs';
  * Resolve `userPath` relative to `allowedRoot` and verify the result
  * stays within `allowedRoot`.
  *
- * - Absolute paths are rejected (they would escape the root).
+ * - Absolute paths that land outside the root are rejected.
+ * - Absolute paths inside the root are allowed (LLM may reference them).
  * - `..` components that climb above the root are rejected.
- * - Symlinks are resolved via `realpathSync` when the target exists.
+ * - Symlinks are resolved via `realpathSync` for existing targets.
+ * - For non-existent targets, the nearest existing ancestor is resolved
+ *   to prevent symlink-parent escapes on write operations.
  *
  * @param {string} userPath  Path supplied by the tool call.
  * @param {string} allowedRoot  Project root directory.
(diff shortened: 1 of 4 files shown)


server/utils/safePath.js
```diff
@@ -0,0 +1,86 @@
...
+  // nearest existing ancestor and verify THAT is inside the root.
+  // This prevents symlink-parent escapes: if repo/link -> /external/,
+  // then repo/link/new.txt should be blocked because realpath(repo/link)
+  // resolves to /external/ which is outside the root.
+  let real;
+  try {
+    real = fs.realpathSync(resolved);
+  } catch {
+    // Target doesn't exist — walk up to the nearest existing ancestor
+    // and resolve it.  This catches symlink escapes at any depth:
```
Zhang-Henry: **Issue: Legitimate symlinks inside the project that point outside the root are blocked**

The `realpathSync` resolution means that if a project has an internal symlink like `data -> /mnt/storage/data` (common for large datasets, shared assets, etc.), then:

```
safePath('data/file.csv', root)
→ realpathSync resolves to /mnt/storage/data/file.csv
→ outside normalizedRoot
→ throws Error
```

This blocks Read, Write, Edit, Glob, Grep, and LS on any symlinked path that points outside the project root. This is a behavioral regression that could break legitimate workflows.

**Suggested fix:** Consider either:
1. An allowlist mechanism (e.g. `SAFEPATH_ALLOWED_ROOTS` env var or a `.safepath.json` config) for trusted external symlink targets.
2. Only check the *logical* resolved path (without `realpathSync`) for `..` traversal and absolute-path escapes, and skip the symlink resolution — this blocks the `../../../etc/passwd` attack without blocking legitimate symlinks. The symlink-parent-escape scenario (`repo/link -> /external/`) is much rarer and arguably less of a threat since it requires the user to have created the symlink themselves.

server/utils/safePath.js
```diff
@@ -0,0 +1,86 @@
...
+ *   to prevent symlink-parent escapes on write operations.
+ *
+ * @param {string} userPath  Path supplied by the tool call.
+ * @param {string} allowedRoot  Project root directory.
+ * @returns {string}  The resolved, validated absolute path.
+ * @throws {Error}  If the path escapes `allowedRoot`.
+ */
+export function safePath(userPath, allowedRoot) {
+  if (!userPath) return allowedRoot;
+
```
Zhang-Henry, line 35: **Minor: Inconsistent return format for falsy input**

When `userPath` is falsy, this returns `allowedRoot` as-is (the raw path). But for all other inputs, the function returns a `realpathSync`-resolved path via `normalizedRoot`.

If the project root itself is a symlink (e.g. `/home/user/project -> /mnt/projects/foo`), then:
- `safePath('', root)` → `/home/user/project` (symlink path)
- `safePath('src/a.js', root)` → `/mnt/projects/foo/src/a.js` (real path)

This could cause subtle path-comparison mismatches downstream. Consider normalizing the early return too:
```js
if (!userPath) {
  try { return fs.realpathSync(allowedRoot); } catch { return allowedRoot; }
}
```

Zhang-Henry (commented):
Good fix for a real vulnerability. The symlink ancestor-walk logic is well thought out. Two issues to consider before merging.

haoyu-haoyu:
Thanks @Zhang-Henry — both issues addressed in 6eeb58f:

**1. Legitimate symlinks blocked → fixed**

Adopted your option 2: `safePath` now performs only a lexical check (`path.resolve` + prefix comparison), removing all `realpathSync` usage and the ancestor-walk logic. This means:

- `data -> /mnt/storage/data` symlinks inside the project **work as expected**
- `../../../etc/passwd` is still **blocked** (path.resolve normalizes away `..`)
- `/etc/passwd` is still **blocked** (absolute path outside root)

The trade-off (symlink-based escapes not caught) is acceptable since creating a malicious symlink requires the Bash tool which needs user approval.

Added a test that creates a symlink to `os.tmpdir()` and verifies it's allowed.

**2. Inconsistent return for falsy input → fixed**

Now returns `path.resolve(allowedRoot)` for falsy input, matching all other code paths. Added a test with a non-normalized root (`ROOT/src/..`) to verify normalization.

The implementation went from 86 lines to 52 lines — much simpler to audit. All 11 tests pass.

PR #138, opened 2026-04-07
feat(chat): add Claude /btw side-question

davidliuk:
## Summary
Adds a Claude Code–style `/btw` side question: one-shot, tool-free answer in an overlay from current chat context, without appending to the main transcript. Only when the chat provider is **Claude**.

## Changes
- `runClaudeBtw` + `POST /api/claude/btw`
- Built-in `/btw` in commands API
- `BtwOverlay` UI; slash commands can run while the model is streaming (textarea not disabled for `/…`)

## Notes
- Other harnesses (Cursor, Codex, Gemini, etc.) still show the in-chat hint to switch to Claude.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dr-claw",
-  "version": "1.1.2",
+  "version": "1.1.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "dr-claw",
-      "version": "1.1.2",
+      "version": "1.1.3",
       "hasInstallScript": true,
       "license": "(GPL-3.0-only AND AGPL-3.0-only)",
       "dependencies": {
@@ -177,7 +177,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -595,7 +594,6 @@
       "resolved": "https://registry.npmjs.org/@codemirror/language/-/language-6.12.2.tgz",
       "integrity": "sha512-jEPmz2nGGDxhRTg3lTpzmIyGKxz3Gp3SJES4b0nAuE5SWQoKdT5GoQ69cwMmFd+wvFUhYirtDTr0/DRHpQAyWg==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@codemirror/state": "^6.0.0",
         "@codemirror/view": "^6.23.0",
@@ -610,7 +608,6 @@
       "resolved": "https://registry.npmjs.org/@codemirror/lint/-/lint-6.9.5.tgz",
       "integrity": "sha512-GElsbU9G7QT9xXhpUg1zWGmftA/7jamh+7+ydKRuT0ORpWS3wOSP0yT1FOlIZa7mIJjpVPipErsyvVqB9cfTFA==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@codemirror/state": "^6.0.0",
         "@codemirror/view": "^6.35.0",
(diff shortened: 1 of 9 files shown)


bbsngg (commented):
## PR Review: `/btw` side-question feature

Overall the idea is solid and the UX is well thought out. A few issues worth addressing before merge:

### 🔴 Critical

**1. Race condition with `process.env` mutation (`server/claude-sdk.js:818-827`)**

```js
const prevStreamTimeout = process.env.CLAUDE_CODE_STREAM_CLOSE_TIMEOUT;
process.env.CLAUDE_CODE_STREAM_CLOSE_TIMEOUT = '120000';
const queryInstance = query({ ... });
// immediately restores env var
```

This is not safe under concurrent requests — two simultaneous `/btw` calls will stomp on each other's env var. The restore also happens before the stream is consumed (`for await`), so the env var may already be reset by the time it's actually needed. Consider passing this as a config option instead of mutating `process.env`.

**2. Textarea `disabled` removed entirely (`ChatComposer.tsx:381`)**

`disabled={isLoading}` was removed from the textarea entirely (along with `disabled:opacity-50` in the className). This means during normal streaming the user can type freely and the textarea is no longer visually dimmed. The intent was to allow `/` commands while streaming, but the fix is too broad — consider keeping `disabled` and only conditionally allowing input when the text starts with `/`.

### 🟡 Should fix

**3. Submit button allows any `/`-prefixed text during streaming (`ChatComposer.tsx:421-424`)**

```tsx
(isLoading && !input.trim().startsWith('/'))
```

This lets the user submit *any* slash command while streaming (e.g. `/clear`, `/rewind`), not just `/btw`. Other commands probably shouldn't be executable mid-stream.

**4. `handleSubmit` guard reordering (`useChatComposerState.ts:922-952`)**

The `isLoading` check was moved after slash command processing. If a non-`/btw` slash command is entered during streaming, it won't be caught by the `/btw` early-return and will fall through past the `isLoading` guard, potentially triggering a normal message send.

**5. Transcript truncation cuts mid-message (`useChatComposerState.ts:197-200`)**

`out.slice(out.length - BTW_TRANSCRIPT_MAX_CHARS)` can cut a message in half. Should find the next `\n\n` boundary after the cut point to preserve message integrity.

**6. No request abort on overlay close**

If the user closes the overlay while the API call is in flight, the fetch continues and the response is silently discarded. Should use `AbortController` to cancel the request on close.

### 💡 Minor

- No server-side size limit on `question` field — a very long question + 120K transcript could create an oversized prompt.
- `BtwOverlay` is rendered inside `ChatComposer` but is semantically unrelated — consider rendering at `ChatInterface` level.
- `CLOSED_BTW_OVERLAY` default shape is duplicated in `closeBtwOverlay` callback and `useState` initializer — could reuse the constant.

### ✅ What looks good

- Clean overlay UI with Escape + backdrop click-to-close
- Proper `aria-modal` and `role="dialog"` accessibility
- Non-Claude providers get a clear fallback message
- Transcript builder correctly filters to user/assistant messages
- Server-side input validation on the route

davidliuk:
All review items addressed in ce85e72. Here's the point-by-point resolution:

### 🔴 Critical

**1. Race condition with `process.env` mutation** — Moved env var restore into a `finally` block so the stream timeout persists through the entire `for await` loop, not just the `query()` constructor call.

**2. Textarea `disabled` removed entirely** — This turned out to be trickier than just restoring `disabled`. Adding `disabled={isLoading && !input.startsWith('/btw')}` creates a catch-22: when streaming starts the input is empty → textarea disabled → user can't type `/btw` at all. Fix: removed the `disabled` attribute, applied conditional `opacity-50` CSS class for visual dimming during streaming. The submit button + `handleSubmit` guard enforce that only `/btw` can be submitted while streaming.

### 🟡 Should fix

**3. Submit button allows any `/`-prefixed text** — Changed from `startsWith('/')` to `startsWith('/btw ')` on the submit button disabled logic.

**4. `handleSubmit` guard reordering** — Added `if (isLoading && commandName !== '/btw') return;` inside the matched-command block, before `executeCommand`. Non-btw slash commands are now blocked during streaming.

**5. Transcript truncation cuts mid-message** — Now finds the next `\n\n` boundary (within 2000 chars) after the cut point before slicing.

**6. No request abort on overlay close** — Added `AbortController` ref; signal passed to `authenticatedFetch`; `closeBtwOverlay` aborts in-flight request; catch block silently returns on abort. Server-side: `runClaudeBtw` accepts a `signal` param, checks `signal.aborted` in the stream loop and returns early. Route uses `res.on('close')` + `writableFinished` guard to detect real client disconnect (not `req.on('close')` which fires prematurely when Express json middleware drains the POST body).

### 💡 Minor

- **Server-side size limit** — Added `MAX_QUESTION_CHARS = 2000` and `MAX_TRANSCRIPT_CHARS = 150_000` with validation in the route.
- **BtwOverlay placement** — Moved from ChatComposer to ChatInterface level.
- **`CLOSED_BTW_OVERLAY` dedup** — Extracted shared constant in `useChatComposerState.ts`, used by both `useState` init and `closeBtwOverlay`.
- **Bonus**: Stabilized `getChatMessagesForBtw` with `useRef` + `useCallback` to avoid unnecessary `executeCommand` re-creation on every `chatMessages` change. Consolidated duplicate `chatMessagesRef`.

PR #155, opened 2026-04-09
Add abortable loading for chat file preview

bbsngg:
## Summary
- add loading-state close actions for chat file previews and full-screen file editor loading
- abort in-flight file content requests when the preview is closed or unmounted
- lazy-load the chat preview editor and show a suspense fallback while it initializes

## Testing
- npm run typecheck
- npm run build

## Notes
- Vite build still reports that `CodeEditor` remains statically imported elsewhere, so the dynamic import in chat does not currently split it into a separate chunk.

### DIFF
--- a/src/components/CodeEditor.jsx
+++ b/src/components/CodeEditor.jsx
@@ -180,6 +180,12 @@ function CodeEditor({ file, onClose, projectPath, selectedProject = null, onStar
   const [candidates, setCandidates] = useState(null);
   const [resolvedPath, setResolvedPath] = useState(null);
   const editorRef = useRef(null);
+  const abortRef = useRef(null);
+
+  const handleAbortAndClose = () => {
+    abortRef.current?.abort();
+    onClose();
+  };
 
   // Check if file is markdown
   const isMarkdownFile = useMemo(() => {
@@ -488,6 +494,10 @@ function CodeEditor({ file, onClose, projectPath, selectedProject = null, onStar
 
   // Load file content
   useEffect(() => {
+    const controller = new AbortController();
+    abortRef.current = controller;
+    const signal = controller.signal;
+
     const loadFileContent = async () => {
       try {
         setLoading(true);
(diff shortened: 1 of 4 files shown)


bbsngg:
Review findings:

1. This change makes the first file-open path slower, not faster. `ChatInterface` now lazy-loads `CodeEditor`, but the actual file read still starts from `CodeEditor`'s `useEffect` after that module mounts. That turns the flow into: click file -> wait for editor module -> start `readFile`/`getFileContentBlob`. For the reported issue (opening files from chat is very slow), this is a regression in the critical path rather than an improvement.

2. The new close button only cancels in-flight file reads after `CodeEditor` has already loaded. The `Suspense` fallback can hide the preview UI, but it cannot cancel the lazy module load itself. So if the visible delay is dominated by editor initialization, this PR still does not let the user close the real loading work; it only closes the shell around it.

Build output also shows Vite warning that `CodeEditor` is still statically imported elsewhere, so the dynamic import here does not currently split it into a separate chunk. That makes the performance benefit even less likely.

Validation run locally:
- `npm run typecheck`
- `npm run build`

bbsngg:
Updated this PR after rebasing onto the latest `main` and addressing the review concern about the chat file-open path.

What changed:
- switched chat file clicks from opening the heavy inline `CodeEditor` directly to opening the lightweight preview modal first
- kept `Open in editor` wired to the existing editor sidebar flow for full editing
- passed `AbortController.signal` through `ChatContextFilePreview` file/blob requests so closing the preview during load now cancels the in-flight request
- kept the existing abortable loading behavior in `CodeEditor`
- removed the chat-specific lazy-editor loading path from the critical open flow

Validation:
- `npm run typecheck`
- `npm run build`

Notes:
- the branch was rebased onto the latest `origin/main` and force-pushed
- local uncommitted `package-lock.json` changes were intentionally kept out of the PR

PR #156, opened 2026-04-09
fix(chat): prevent right sidebar overflow in narrow windows

MarkSiqiZhang:
## Summary
   - Remove `flex-shrink-0` from expanded sidebar so it shrinks naturally within the flex container instead of overflowing the viewport
   - Add horizontal scroll to tab bar when sidebar is too narrow to show all tabs
   - Ensure the collapse button always stays visible with `flex-shrink-0`
   - Cap drag-resize width to available space so sidebar can't be dragged wider than the window allows

## Test plan
   - [ ] Shrink browser window horizontally — sidebar should shrink with it, no overflow
   - [ ] Collapse button remains visible and functional at all widths
   - [ ] Tab bar scrolls horizontally when too narrow for all tabs
   - [ ] Drag-resize sidebar in a narrow window — should cap at available space
   - [ ] Expand window back — sidebar returns to its configured width
   - [ ] Mobile layout unaffected

### DIFF
--- a/src/components/chat/view/subcomponents/ChatContextSidebar.tsx
+++ b/src/components/chat/view/subcomponents/ChatContextSidebar.tsx
@@ -57,6 +57,7 @@ const SIDEBAR_SECTIONS_STORAGE_KEY = 'chat-session-context-sections';
 const DEFAULT_SIDEBAR_WIDTH = 480;
 const MIN_SIDEBAR_WIDTH = 360;
 const MAX_SIDEBAR_WIDTH = 840;
+const MIN_CHAT_AREA_WIDTH = 400;
 const SECTION_STYLES: Record<SectionTone, {
   panel: string;
   glow: string;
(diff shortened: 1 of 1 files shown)


bbsngg:
I reviewed the sidebar overflow fix and don't see any blocking issues in the implementation.

The change makes sense end-to-end: the expanded sidebar can now shrink within the flex layout, the tab strip falls back to horizontal scrolling instead of compressing the buttons, the collapse control is protected from shrinking away, and drag-resize is capped by the actual available container width rather than only the global max width.

The main remaining risk is UI regression coverage rather than code correctness. I'd recommend manually verifying a few edge cases before merge:
- very narrow desktop widths, especially around the sidebar minimum width
- dragging the sidebar to its max in a narrow window, then expanding the window again
- tab bar behavior with longer localized labels to confirm horizontal scrolling still feels usable

From the code side, this looks good to me.

PR #162, opened 2026-04-11
feat(chat): upstream-safe queue/steer + session lifecycle hardening

everett4320:
Scope (public upstream branch only):\n- Port queue/steer workflow for chat dispatch, including queued-turn persistence, steer promotion, pause/resume semantics, and settlement reconciliation.\n- Add session scope isolation helpers and tests to prevent cross-project/provider session collisions when session IDs overlap.\n- Strengthen realtime protocol with explicit lifecycle events: session-accepted, session-busy, and session-state-changed.\n- Enrich lifecycle/session-created payloads with projectName metadata for more reliable client-side routing and state ownership.\n- Keep Nano provider command/session status flows aligned with other providers in active-session and lifecycle handling.\n\nUpstream safety constraints applied:\n- Excluded local/private product policy changes (Codex-only strategy, auth/license stack, workspace-root hard-cut/backfill behavior).\n- Excluded branding and private demo/link routing changes (kept Dr. Claw naming).\n- Avoided introducing server-side CODEX_ONLY/provider lock-in behavior into upstream.\n\nCompatibility notes:\n- Lifecycle protocol additions are provider-agnostic and transport-level; no Windows-only behavior was hardcoded.\n- Existing macOS/Linux provider command paths remain unchanged.\n\nValidation:\n- node --check server/index.js\n- node --check server/claude-sdk.js\n- node --check server/cursor-cli.js\n- node --check server/gemini-cli.js\n- npm run test -- ... (blocked: vitest executable missing in this environment)\n- npm run typecheck (blocked: tsc executable missing in this environment)


Zhang-Henry:
## Code Review

Thanks for the upstream-safe split. Reviewed the full diff (29 files, +6014/−2815). A few issues worth addressing before merge.

### 🔴 Critical

1. **`server/cursor-cli.js` still emits `type: 'claude-complete'` on exit.**
   This PR adds lifecycle event inference that depends on correct provider identification, but the cursor exit path mislabels itself as `claude`. The result: `session-state-changed` carries the wrong provider, and the new session-scope isolation can be bypassed when a real claude session is also active. This is a one-line fix but a hard blocker for the lifecycle work this PR introduces.

2. **No tests for `reconcileSessionQueueId()` / `reconcileSettledSessionQueue()`** in `src/components/chat/utils/__tests__/codexQueue.test.ts`.
   These functions migrate queued turns from temporary session IDs (`new-session-<ts>`) to real session IDs once the session is created — they're the core of the queue/steer flow. Right now there's zero coverage, so ordering bugs, lost turns, or double-dispatch on reconciliation would silently slip through. Please add at least:
   - migration from temp ID → real ID preserves order
   - empty source queue is a no-op
   - non-temporary IDs are not touched

3. **`useChatComposerState.ts` around line 1272 — `resolveSessionContext` dependency array is missing `pathname`.**
   Internally it calls `getRouteSessionId()` which reads `window.location.pathname`, but pathname is not tracked. If the user navigates to a different session URL without `selectedProject` / `currentSessionId` changing, the composer keeps the stale route session ID and can dispatch turns into the wrong session. Either include pathname in the deps or compute the route session ID inside an effect that listens to navigation.

### 🟡 Notable

4. `sendSessionAccepted()` in `server/index.js` (~L1641) sends `session-accepted` immediately followed by `session-state-changed=running`. Correct today because both writes are synchronous, but the client now hard-depends on ordered delivery — worth a comment or a single combined event.

5. The new lifecycle payloads enrich `projectName` via `encodeProjectPath(projectPath)` without verifying the project actually exists. The trust boundary moves entirely upstream onto whatever validates `projectPath` on the request. Worth an explicit assertion at the lifecycle emission site.

6. `src/utils/__tests__/sessionScope.test.ts` only covers cross-project rejection. Please also cover:
   - same `sessionId` across different providers
   - `sessionId` containing the `::` separator
   - empty/undefined scope fields

7. `useChatRealtimeHandlers.ts` `resolveProvider()` silently falls back to the default provider for unknown values (via `normalizeProvider`). It's safe in isolation, but it would have hidden the cursor mislabeling above. Consider logging a warning on unknown providers so similar bugs surface in CI.

### 🟢 Minor / nits

- Several `console.log` debug statements left in: `server/index.js:1701`, `useChatRealtimeHandlers.ts:1234`, `useChatRealtimeHandlers.ts:1327`. Please gate behind a debug flag or drop.
- `resolveProjectName()`'s try/catch around `encodeProjectPath` swallows errors silently — at least log at debug level so failures are diagnosable.
- UI removes the nano provider from `getModelsByProvider`, but the server still handles `nano-command`. Looks intentional, but please call it out in CHANGELOG so downstream isn't surprised.
- `session-accepted` messages include a `projectPath` field that I can't find any client consumer for — looks like a dead field, please remove or document.

### Overall

The lifecycle protocol and scope isolation design are clean and the upstream-safe scoping is well done. The main risk is correctness: the new infrastructure assumes accurate provider identification, and the existing cursor-cli bug breaks that assumption on the cursor path. Fixing #1, adding the reconciliation tests in #2, and patching the `resolveSessionContext` deps in #3 would be the minimum I'd want before merge.

everett4320:
reopen another one

Issue #172, opened 2026-04-13
[Feature]: support SSH profile alias for remote compute node


kinianlo:
### Summary

Currently remote compute nodes are specified by a form with fields like hostname, username, port, etc. There is not a way to allow the user to simply specify a ssh aliases, i.e. the ones defined in ~/.ssh/config. Imagine a user who used to use `ssh machine_alias` to connect to their remote machine. 

### Problem to solve

1. User unable to configure a remote compute node by ssh aliases.
2. User unable to specify a ssh jump host.

### Proposed solution

Introduce an option to the UI which allows the user to fill in the ssh alias. 

### Impact

Affected: users
Severity: blocks workflow
Frequency: always
Consequence: unable to leverage remote machine which requires a jump host to access

Zailushang211:
点test connection出现Error: File not found: ，有人遇到过这个问题吗

zuijiuhanxuan:
> 点test connection出现Error： File not found： ，有人遇到过这个问题吗

遇到了，请问你后面有解决吗？我一直没找到解决方法

PR #180, opened 2026-04-15
feat: SSH agent auth for compute nodes + fix config path

liuyixin-louis:
## Summary
- Add **SSH Agent** as a third authentication method for compute nodes (alongside SSH Key and Password), supporting 1Password, ssh-agent, and other key agents
- Fix `execSsh()`/`execRsync()` to fall back to SSH agent instead of throwing "No authentication method configured" when no explicit key or password is set
- Move config directory from `~/.openclaw` to `~/.dr-claw` for consistency with the rest of the app
- Update `NodeForm` UI with SSH Agent button and explanatory hint text

## Test plan
- [ ] Add a compute node with "SSH Agent" auth — verify no error on Test Connection
- [ ] Existing nodes with SSH Key or Password auth still work
- [ ] New nodes default to "SSH Agent" in the form
- [ ] `~/.dr-claw/compute-node.json` used instead of `~/.openclaw/`
- [ ] Community tools config also writes to `~/.dr-claw/`

### DIFF
--- a/server/compute-node.js
+++ b/server/compute-node.js
@@ -7,6 +7,7 @@ import crypto from 'crypto';
 
 const CONFIG_DIR = path.join(os.homedir(), '.dr-claw');
 const CONFIG_FILE = path.join(CONFIG_DIR, 'compute-node.json');
+const LEGACY_CONFIG_DIR = path.join(os.homedir(), '.openclaw');
 
 // ─── ID generation ───
 
@@ -23,7 +24,19 @@ async function loadRawConfig() {
     const data = await fs.readFile(CONFIG_FILE, 'utf8');
     return JSON.parse(data);
   } catch (e) {
-    return { nodes: [], activeNodeId: null };
+    // Migrate from legacy ~/.openclaw/ if it exists
+    const legacyFile = path.join(LEGACY_CONFIG_DIR, 'compute-node.json');
+    try {
+      const legacyData = await fs.readFile(legacyFile, 'utf8');
+      const parsed = JSON.parse(legacyData);
+      // Copy to new location
+      await fs.mkdir(CONFIG_DIR, { recursive: true });
+      await fs.writeFile(CONFIG_FILE, legacyData, { mode: 0o600 });
+      console.log('[compute-node] Migrated config from ~/.openclaw/ to ~/.dr-claw/');
+      return parsed;
+    } catch {
+      return { nodes: [], activeNodeId: null };
+    }
   }
 }
 
--- a/server/compute-node.js
+++ b/server/compute-node.js
@@ -5,7 +5,7 @@ import os from 'os';
 import pty from 'node-pty';
 import crypto from 'crypto';
 
-const CONFIG_DIR = path.join(os.homedir(), '.openclaw');
+const CONFIG_DIR = path.join(os.homedir(), '.dr-claw');
 const CONFIG_FILE = path.join(CONFIG_DIR, 'compute-node.json');
 
 // ─── ID generation ───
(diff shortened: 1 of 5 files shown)


Zhang-Henry:
## Review notes

Tested CI is clean and the logic changes (SSH agent fallback in `execSsh`/`execRsync`, new auth-type UI) look correct. One concern before merging:

### Missing migration: `~/.openclaw` → `~/.dr-claw`

Context: on `main` the rest of the server already writes to `~/.dr-claw` (`load-env.js`, `openrouter.js`, `local-gpu.js`, `projects.js`, `references.js`, `nanoSessionPaths.js`, …). Only `server/compute-node.js` and `server/routes/community-tools.js` were still writing to `~/.openclaw`, so this PR is fixing a real inconsistency — good.

However, there is no one-time migration. Existing users who already configured compute nodes or community-tools API keys with an older build have their data in `~/.openclaw/compute-node.json` and `~/.openclaw/community-tools.json`. After this PR lands, those files are silently ignored and they’ll see "no nodes configured" / API keys missing.

Suggested fix (either):
- On server startup, if `~/.dr-claw/compute-node.json` / `~/.dr-claw/community-tools.json` don’t exist but the `~/.openclaw/*` counterpart does, copy or rename them (one-shot).
- Or call this out explicitly in release notes.

### Minor

- `GET /config` now returns `authType: 'agent'` when a node has neither `keyPath` nor `password`. That changes the reported type for historical incomplete records — intentional, just worth noting.
- `configured` now only requires `host` + `user` (no credentials). Reasonable since agent is always an implicit fallback.

The `execSsh`/`execRsync` branch rewrites are logically equivalent for the existing key/password cases, confirmed by walking through each `(keyPath?, password?)` combination.

liuyixin-louis:
Thanks for the review! Addressed the migration concern in 87a2f7b:

**Migration logic added:**
- `compute-node.js`: `loadRawConfig()` now checks for `~/.openclaw/compute-node.json` when `~/.dr-claw/compute-node.json` doesn't exist — copies it over on first load (one-shot).
- `community-tools.js`: Same pattern — migrates `~/.openclaw/community-tools.json` to `~/.dr-claw/` on first read if the new path is missing.

Both log when migration happens. Tested three scenarios: legacy-only → migrates, new-path-exists → no-op, neither-exists → empty default.

Re: the minor notes — yes, `authType: 'agent'` for historical records without credentials and `configured` only requiring host+user are both intentional.

server/compute-node.js
```diff
@@ -23,7 +24,19 @@ async function loadRawConfig() {
...
   } catch (e) {
-    return { nodes: [], activeNodeId: null };
+    // Migrate from legacy ~/.openclaw/ if it exists
+    const legacyFile = path.join(LEGACY_CONFIG_DIR, 'compute-node.json');
+    try {
+      const legacyData = await fs.readFile(legacyFile, 'utf8');
+      const parsed = JSON.parse(legacyData);
+      // Copy to new location
+      await fs.mkdir(CONFIG_DIR, { recursive: true });
+      await fs.writeFile(CONFIG_FILE, legacyData, { mode: 0o600 });
```
davidliuk, line 34: [P2] Migrate only when the current config is absent, not on any read/parse failure

The outer catch includes JSON parse errors and all read errors. A malformed existing `~/.dr-claw/compute-node.json` therefore triggers this write and is replaced with stale `~/.openclaw/compute-node.json`. I reproduced this in a temporary HOME: a current file containing a newer node but truncated JSON was overwritten during `loadAllNodes()` with the legacy node list, without backup or an error. Limit migration to ENOENT, preserve/report malformed or unreadable current files, and avoid clobbering a concurrently-created destination. The new community-tools migration catch has the same parse-error fallback pattern. Please test that an existing malformed destination remains byte-for-byte untouched.

davidliuk (commented):
Reviewed the current head and previous migration discussion. The SSH-agent fallback and auth selector are useful; migration is now present, so I am not repeating the earlier missing-migration finding. I reproduced a separate overwrite-on-parse-error issue in a temporary HOME (inline).

For issue triage, this is **not yet a complete fix for #172**. The create-node API still requires a username, defaults port to 22, and SSH/rsync always pass explicit `user@host` and `-p`. Entering only a profile alias cannot inherit its User/Port as `ssh machine_alias` does. OpenSSH can already inherit ProxyJump from a matching Host entry, but there is still no alias-only configuration mode here. Please keep #172 open and track that remaining work separately; add an alias-mode regression with HostName/User/Port/ProxyJump and local `ssh -G` before claiming coverage. No live SSH connection or credentials were used.

PR #183, opened 2026-04-15
refactor: simplify file preview and clean up session context summary

bbsngg:
## Summary
- Remove JSON syntax highlighting from ChatContextFilePreview (use plain text for non-markdown/html/media)
- Switch skill preview in ChatContextSidebar to global skills tree API
- Remove ~500 lines of unused path-parsing regexes and hardcoded file extension lists from sessionContextSummary

Part of the session lifecycle PR split (1/6, independent).

## Test plan
- [ ] File preview renders correctly for text, markdown, image, PDF, audio, video files
- [ ] Skill preview in context sidebar loads via global skills API
- [ ] `npx vitest run src/components/chat/utils/__tests__/sessionContextSummary.test.ts` passes


bbsngg:
Split from #170 (1/6). Independent — can be merged immediately. See #170 for full split plan.

bbsngg (commented):
## Code Review

Overall a solid cleanup — removes ~678 net lines of fragile shell path-parsing logic, and the removals are safely scoped (all deleted functions/constants are internal-only). A few items to address before merge:

---

### P0 — `sandbox="allow-scripts allow-same-origin"` (ChatContextFilePreview.tsx)

`sandbox=""` → `sandbox="allow-scripts allow-same-origin"` effectively neutralizes iframe sandboxing. Scripts running with same-origin access can programmatically remove the sandbox attribute, meaning a malicious HTML file could access the parent page's cookies, localStorage, and make authenticated API calls.

`SurveyPage.tsx:244` already uses the same combo, so it's at least consistent — but the old `sandbox=""` was strictly more secure. Consider whether HTML previews actually need scripts. If not, keep `sandbox=""` or use only `sandbox="allow-same-origin"`.

---

### P1 — Inlined extension sets diverge from shared `fileExtensions.ts`

The inlined sets in `ChatContextFilePreview.tsx` are **missing entries** compared to the shared `fileExtensions.ts`:

| Set | Missing | Present in `fileExtensions.ts` |
|---|---|---|
| `AUDIO_EXTENSIONS` | `aac`, `opus` | ✅ |
| `VIDEO_EXTENSIONS` | `flv`, `avi` | ✅ |

These file types will render as plain text instead of media players. `FileThumbnail.tsx` still imports from `fileExtensions.ts`, so the shared module must remain and the duplication means future additions need two updates.

Recommendation: keep importing from `fileExtensions.ts`, or explicitly document the subset.

---

### P1 — Performance: full skills tree fetched on every click

`ChatContextSidebar.tsx` now calls `api.getGlobalSkills()` (full tree) + BFS + `api.readGlobalSkillFile()` on **every skill preview click** — two round-trips with no caching, replacing a single `api.resolveSkill()` call.

Consider caching the tree response in a ref, or keeping `resolveSkill` for this path.

---

### P2 — `api.resolveSkill` left dangling

After this PR, `api.resolveSkill` (`api.js:160`) has zero callers. Remove it here or track for cleanup.

### P2 — AbortController removed

Replaced with a `cancelled` boolean. This prevents stale state updates but doesn't cancel in-flight network requests. Both `api.readFile` and `api.getFileContentBlob` accept `{ signal }`. Rapid file switching will leave orphaned fetches.

### P2 — `case 'Skill'` tool name dropped

Only `activate_skill` is handled now. If any provider emits `toolName: 'Skill'`, those skills will be silently dropped from the session summary.

---

### P3 — Minor behavior changes (just flagging)

- **`UpdatePlan` kind**: changed from `'todo'` to `'task'` — may affect sidebar grouping
- **`.txt` files**: no longer render as markdown (now plain `<pre>`) — probably correct
- **Text preview styling**: `whitespace-pre-wrap break-words` removed — long lines will scroll horizontally instead of wrapping

---

### What looks good

- All removed code is safely scoped — no external callers
- `case 'update_plan'` removal is safe (server normalizes to `UpdatePlan` in `codexSessionMessages.js`)
- Test removals match code removals cleanly
- `extractSkillContext` returning `{ label, path? }` is cleaner than the old separate-extract-then-reparse approach

PR #191, opened 2026-04-16
fix(chat): render LaTeX \\[..\\] / \\(..\\) delimiters (Issue #50)

### LINKED ISSUE #50: [Feature]: Markdown rendering issues: Mermaid / LaTeX / preformatted ASCII blocks are shown as raw or broken text
### Summary

Some generated outputs are not rendered correctly in the UI. Instead of being displayed as diagrams, math, or properly formatted preformatted blocks, they appear as raw text or broken layouts.

### Problem to solve

This affects at least these cases:

1. **Pipeline / diagram-like content**
   - Mermaid-like or boxed pipeline content is partially rendered as broken horizontal lines / raw text.
   - Fence markers or language hints such as `text` seem to leak into the rendered output.
   - Layout is fragmented across multiple blocks.

2. **LaTeX math content**
   - Block math is shown literally, e.g. `\[ ... \]`, `\lambda`, `\mathrm`, etc.
   - Equations are not being rendered into formatted math.

3. **ASCII / monospace comparison blocks**
   - Alignment is broken.
   - Vertical bars and separators wrap incorrectly.
   - The result becomes hard to read.

### 1) "DCA 三阶段 Pipeline"
Expected:
- A clean structured diagram / preformatted block

Actual:
- Broken lines appear above the main block
- Raw `text` / fence-related content leaks into the UI
- Some sections are split into separate dark blocks unexpectedly

### 2) "单个子问题长什么样"
Expected:
- Properly rendered math formula

Actual:
- Raw LaTeX source is displayed directly, including `\[...\]` and escaped commands

### 3) "速度对比结果"
Expected:
- A readable structured comparison block / table / diagram

## Attachments

I attached screenshots showing:
- broken pipeline rendering
- raw LaTeX rendering
- broken speed comparison block

Image

Image

Image

### Proposed solution

-

### Alternatives considered

_No response_

### Impact

-

### Evidence/examples

_No response_

### Additional information

_No response_

MarkZhang777:
## Summary                                                                                                                                    
   - Fixes #50: chat messages containing LaTeX `\[..\]` (display) / `\(..\)` (inline) now render via KaTeX instead of showing as raw source.     
   - Eliminates a connected `\theta → <TAB>heta` corruption that hit bracket-delimited blocks because the existing math mask only covered `$..$` / `$$..$$`.
   - Chat surface only. `ResearchLab.jsx`, `CodeEditor.jsx`, `ChatContextFilePreview.tsx`, `SurveyPage.tsx` don't route through `unescapeWithMathProtection` and are left for a follow-up.

   ## Root cause
   Two bugs stacked:
   1. `remark-math@6` recognizes only `$` delimiters, so `\[..\]` / `\(..\)` never became math nodes. Worse, CommonMark treats `\[` / `\]` / `\(` / `\)` as valid backslash-escapes of ASCII punctuation, so remark silently stripped the backslashes, leaving `[ r_t(...) ]` rendered as prose.
   2. The existing `unescapeWithMathProtection` protects math blocks from its own `\\t → \t` replacement via a mask/unmask trick — but the mask regex only matched `$..$` / `$$..$$`. Any `\theta` / `\tau` / `\n*`-command **inside** bracket-delimited math lost the backslash when its `\t` was turned into a literal TAB character.

## Changes
   | File | Purpose |
   |---|---|
   | `src/utils/latexNormalizer.ts` (new) | Rewrites `\[..\]` → `$$..$$` and `\(..\)` → `$..$` outside fenced code blocks and inline code. Single-pass alternation regex, fast-path short-circuit when no brackets present, returns same string reference on the no-op path (preserves `useMemo` identity). |
   | `src/components/chat/utils/chatFormatting.ts` | Calls the normalizer at the top of `unescapeWithMathProtection`, so bracket math is rewritten to `$`-form **before** the existing mask/replace/unmask runs — bringing it under the same protection umbrella. |

## Why hook into `unescapeWithMathProtection` instead of the 5 ReactMarkdown call sites
   All 10+ chat render paths already funnel through `unescapeWithMathProtection` (via `messageTransforms.ts` and  `useChatRealtimeHandlers.ts`). A 2-line patch there covers the whole chat surface without touching any React component. The non-chat renderers don't call this function and are deferred to a follow-up PR.

## Test plan
   - [ ] Ask the assistant to emit a display formula explicitly using `\[ r_t(\theta) = \frac{\pi_\theta(a|s)}{\pi_{\theta_{old}}(a|s)} \]` — verify KaTeX renders the fraction (DOM check: `<span class="katex">` present, with both `katex-mathml` and `katex-html` children).
   - [ ] Ask for inline math with `\( y = mx + b \)` — verify KaTeX inline rendering.
   - [ ] Verify existing `$..$` / `$$..$$` messages still render (no regression).
   - [ ] Send a fenced code block containing `arr\[0\]` (e.g. a Rust sample) — verify the code is NOT mangled into `arr$$0$$`.
   - [ ] Verify `\theta` / `\tau` / `\pi_{new}` inside bracket-delimited math are intact (no `<TAB>heta` corruption).
   - [ ] `npm run typecheck` — passes
   - [ ] `npm run build` — passes

### DIFF
--- a/src/utils/latexNormalizer.ts
+++ b/src/utils/latexNormalizer.ts
@@ -1,0 +1,29 @@
+// remark-math only recognizes $-delimiters; rewrite \[..\] and \(..\) so KaTeX
+// sees them. Fenced code blocks and inline code pass through untouched.
+
+const PROTECTED_SEGMENTS = /```[\s\S]*?```|`[^`\n]*`/g;
+const MATH_DELIMITERS = /\\\[([\s\S]+?)\\\]|\\\(([^\n]+?)\\\)/g;
+
+export function normalizeLatexDelimiters(input: string): string {
+  if (input.length === 0) return input;
+  if (!input.includes('\\[') && !input.includes('\\(')) return input;
+
+  let output = '';
+  let cursor = 0;
+
+  for (const match of input.matchAll(PROTECTED_SEGMENTS)) {
+    const start = match.index;
+    output += rewriteMathDelimiters(input.slice(cursor, start));
+    output += match[0];
+    cursor = start + match[0].length;
+  }
+  output += rewriteMathDelimiters(input.slice(cursor));
+  return output;
+}
+
+function rewriteMathDelimiters(segment: string): string {
+  if (!segment) return segment;
+  return segment.replace(MATH_DELIMITERS, (_, display, inline) =>
+    display !== undefined ? `$$${display}$$` : `$${inline}$`,
+  );
+}
--- a/src/components/CodeEditor.jsx
+++ b/src/components/CodeEditor.jsx
@@ -17,6 +17,7 @@ import remarkGfm from 'remark-gfm';
 import remarkMath from 'remark-math';
 import rehypeKatex from 'rehype-katex';
 import rehypeRaw from 'rehype-raw';
+import { normalizeLatexDelimiters } from '../utils/latexNormalizer';
 import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';
 import { oneDark as prismOneDark } from 'react-syntax-highlighter/dist/esm/styles/prism';
 import { api } from '../utils/api';
@@ -138,7 +139,7 @@ function MarkdownPreview({ content }) {
       rehypePlugins={rehypePlugins}
       components={markdownPreviewComponents}
     >
-      {content}
+      {normalizeLatexDelimiters(content ?? '')}
     </ReactMarkdown>
   );
 }
--- a/src/components/ResearchLab.jsx
+++ b/src/components/ResearchLab.jsx
@@ -4,6 +4,7 @@ import ReactMarkdown from 'react-markdown';
 import remarkGfm from 'remark-gfm';
 import remarkMath from 'remark-math';
 import rehypeKatex from 'rehype-katex';
+import { normalizeLatexDelimiters } from '../utils/latexNormalizer';
 import {
   FlaskConical, RefreshCw, FileText, BookOpen, Settings2, Lightbulb,
   GitBranch, FolderOpen, ChevronDown, ChevronRight, ExternalLink,
@@ -755,7 +756,7 @@ function OverviewCard({ instance, config, researchBrief, compact = false }) {
                           rehypePlugins={[rehypeKatex]}
                           components={OVERVIEW_MARKDOWN_COMPONENTS}
                         >
-                          {section.content}
+                          {normalizeLatexDelimiters(section.content)}
                         </ReactMarkdown>
                       </div>
                     </div>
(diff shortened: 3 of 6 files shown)


src/utils/latexNormalizer.ts
```diff
@@ -0,0 +1,29 @@
+// remark-math only recognizes $-delimiters; rewrite \[..\] and \(..\) so KaTeX
+// sees them. Fenced code blocks and inline code pass through untouched.
+
+const PROTECTED_SEGMENTS = /```[\s\S]*?```|`[^`\n]*`/g;
```
davidliuk, line 4: [P2] Preserve all Markdown code regions, including unfinished streamed fences

This regex protects only closed triple-backtick fences and single-backtick spans. Reproduced inputs:
- `~~~latex\n\[ x^2 \]\n~~~`
- a four-space-indented `\[ x^2 \]`
- an unfinished streamed fence: opening triple backticks + `latex`, newline, `\[ x^2 \]` (no closing fence yet)

Each literal source is rewritten to `$$ x^2 $$`. The unfinished case affects chat while tokens arrive; tilde/indented blocks remain corrupted after completion. Please use Markdown-aware code-region detection (or track fence type/length and protect through EOF), with unchanged-source regressions for these cases.

src/utils/latexNormalizer.ts
```diff
@@ -0,0 +1,29 @@
...
+    cursor = start + match[0].length;
+  }
+  output += rewriteMathDelimiters(input.slice(cursor));
+  return output;
+}
+
+function rewriteMathDelimiters(segment: string): string {
+  if (!segment) return segment;
+  return segment.replace(MATH_DELIMITERS, (_, display, inline) =>
+    display !== undefined ? `$$${display}$$` : `$${inline}$`,
```
davidliuk, line 27: [P2] Emit block math for the display delimiter branch

For the common single-line display formula `\[ x^2 \]`, this emits `$$ x^2 $$`. Using the actual remark-math/rehype-katex pipeline, I get `<p><span class="katex">…</span></p>` and **no `katex-display`**: same-line double dollars are parsed as inline math, so the result is not a display equation. Please emit an unambiguous math block (delimiter lines/body separated by newlines, respecting paragraph boundaries) or create a math AST node. The regression should assert display-mode output, not merely the presence of `.katex`.

davidliuk (commented):
Reviewed the current head, including the preview-surface changes (the description still says chat-only, but the current diff now covers those previews too). I reproduced two gaps using the actual normalizer and ReactMarkdown + remark-math + rehype-katex in an isolated checkout; see inline comments. Ordinary bracket math reaches KaTeX, but code preservation and display-mode semantics need regression coverage before treating the math part of #50 as fully resolved. No working-project files or user configuration were changed.

PR #192, opened 2026-04-17
feat(chat): one-click "Revert changes" button for agent modifications (closes #85)

### LINKED ISSUE #85: [Feature]: Add one-click "Revert/Discard Changes" for Coding Agent modifications
### Summary

A one-click UI button to instantly discard and revert all file modifications generated by a specific Coding Agent output.

### Problem to solve

When iterating on complex experimental workflows or prototyping computer vision architectures, I heavily rely on the Dr. Claw Coding Agent to write and modify code. However, when the agent hallucinates, introduces bugs (such as incorrect tensor dimensions), or takes the logic in the wrong direction, manually untangling those changes completely breaks the development flow. Currently, resetting the workspace requires context-switching to the terminal to use Git commands (like git restore .), which disrupts the rapid trial-and-error process essential for AI research.

### Proposed solution

A straightforward, one-click "Discard Changes" or "Undo" button located directly within the Dr. Claw UI, ideally attached to the Coding Agent's specific output or diff view. Triggering this should instantly revert all files modified during that specific agent interaction back to their exact prior state, safely rolling back the unwanted code.

### Alternatives considered

Manual Undo: Using standard global IDE undo (Ctrl+Z / Cmd+Z), which is extremely error-prone and unreliable when the agent edits multiple files simultaneously.

Git Commands: Executing git restore . or git checkout in the terminal. This forces a context switch out of the Dr. Claw environment and risks losing other uncommitted, manual script adjustments I might want to keep.

### Impact

Affected users/systems/channels: All Dr. Claw users relying on the Coding Agent for rapid code iteration and experimental prototyping (such as testing new computer vision architectures or tuning training scripts).
Severity: Blocks workflow / Highly annoying. It severely breaks the developer's "flow state" by forcing manual intervention to clean up unwanted code.
Frequency: Frequent. It occurs whenever the agent misunderstands a complex prompt, hallucinates incorrect logic, or modifies tensor dimensions incorrectly across multiple files.
Consequence: Causes significant delays, forces context-switching (e.g., dropping into the terminal to run git restore), and introduces a high risk of accidentally losing manual, uncommitted code while trying to untangle the agent's flawed edits.

### Evidence/examples

_No response_

### Additional information

Since Dr. Claw is designed as a comprehensive AI research IDE, users frequently prompt the agent with highly experimental tasks. Making it completely safe and effortless to "roll back" a failed agent experiment will significantly enhance the coding experience. It empowers users to attempt more ambitious prototyping without the constant fear of the agent corrupting the current working state.

haoyu-haoyu:
## Summary

Adds a **"Revert changes"** button to each finished Coding Agent turn so users can roll back that turn's file modifications from the chat UI instead of context-switching to the terminal for `git restore .`.  Closes #85.

## What it does

- Scans the messages in the finished agent turn for `Edit` / `Write` / `ApplyPatch` / `MultiEdit` / `FileChanges` tool calls
- Extracts every file path those tools actually modified (handles both object and stringified-JSON `toolInput`, multi-file `file_paths` arrays, Codex's `kind: path` newline format, and object-map `changes` from patch-apply results)
- Click → inline confirmation with an explicit warning → `POST /api/git/revert-agent-changes`
- Backend reverts each file independently:
  - untracked regular files → unlink (directories and out-of-repo symlinks are refused)
  - tracked worktree+staged modifications → `git restore --source=HEAD --staged --worktree` (handles `MM` / `AM` / `AD` / `MD` / `DM`)
  - staged adds → `git reset HEAD` + unlink
  - rename / copy pairs → skipped (refuses to leave half-renamed state)
  - per-file error reporting so one bad entry doesn't abort the rest

## Safety

- **Only touches files this specific agent turn modified** — the main ask from the issue.  Any manual edits to unrelated files are preserved.
- Path traversal blocked via the existing `safePath` helper.
- Untracked-symlink deletion re-validates `realpath()` against the project root before `unlink`.
- `git status --porcelain -z` prevents filenames containing `" -> "` from confusing rename parsing.
- Confirmation dialog explicitly warns: *"Any manual edits to these files will also be lost"* — honest about `git restore` semantics since we restore to `HEAD`, not a pre-agent snapshot.  Per-turn exact-snapshot revert (with conflict detection for files the user also touched) would need pre-agent content capture and feels like a separate PR.

## Accessibility

- `role="group"` on the inline confirmation with `aria-label`, focus moves to the destructive action on open, Escape cancels and restores focus to the trigger
- `role="status"` + `aria-live` on reverting / done / error states
- Decorative SVGs marked `aria-hidden`

## Test plan

All 130 Vitest tests pass (incl. 8 new backend integration tests against a real temp git repo, 1 in-process Express route test, and 12 new frontend unit tests).  `tsc --noEmit` clean.

- [x] `npm test -- --run` (130 passed, 18 test files)
- [x] `npm run typecheck` (clean)
- [x] Backend: untracked file delete, tracked file restore, staged add, mixed `MM` state, rename skip, rename-with-modification skip, symlink safety (external file preserved), directory rejection, path traversal blocked, filename containing `" -> "`
- [x] Frontend: object `toolInput`, stringified-JSON `toolInput`, multi-file `file_paths`, Codex `FileChanges` string format, object-map `changes`, array-of-strings `changes`, malformed JSON, subagent children, tool-error skipping, no-result skipping
- [x] Route: 400 on missing `project`, 400 on non-array `files`, empty list, happy path, partial success reporting
- [ ] Manual smoke test: run agent, modify multiple files, click Revert, confirm only those files roll back

## i18n

Added `revertAgent` block to `en`, `zh-CN`, `ko` chat locales with i18next pluralization (`_one` / `_other`).

## Follow-ups (not in this PR)

- Per-turn exact-snapshot revert with conflict detection (requires capturing pre-agent content)
- Rename handling: currently skipped; could surface a custom UI prompt
- Korean locale has some unrelated pre-existing missing keys (`providerInfo.nano`, `readyPrompt.nano`) — left for a separate i18n-sync PR

### DIFF
--- a/server/routes/git.js
+++ b/server/routes/git.js
@@ -5,6 +5,7 @@ import { promises as fs } from 'fs';
 import { extractProjectDirectory } from '../projects.js';
 import { queryClaudeSDK } from '../claude-sdk.js';
 import { spawnCursor } from '../cursor-cli.js';
+import { safePath } from '../utils/safePath.js';
 
 const router = express.Router();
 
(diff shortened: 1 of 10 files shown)


server/routes/git.js
```diff
@@ -1343,4 +1344,217 @@ router.post('/delete-untracked', async (req, res) => {
...
+          } catch (unlinkError) {
+            if (unlinkError.code !== 'ENOENT') throw unlinkError;
+          }
+        } else if ('MDT'.includes(indexChar) || 'MDT'.includes(worktreeChar) || status === 'AD') {
+          // Anything that existed at HEAD and has staged and/or worktree
+          // modifications, deletions, or type changes.  --staged --worktree
+          // together restore BOTH the index and the working tree to HEAD
+          // so mixed states like MM / AM / AD / MD / DM all land back
+          // at the committed state.
+          await spawnAsync('git', ['restore', '--source=HEAD', '--staged', '--worktree', '--', relFile], {
```
davidliuk, line 1521: [P1] Treat selected filenames as literal Git pathspecs

`--` stops option parsing but does not disable Git pathspec glob/magic expansion. In a temporary repo I committed both `agent*.txt` (a literal filename) and `agent-user.txt`, modified both, then called `revertFilesAtProjectPath(repo, ['agent*.txt'])`. The result reported only `agent*.txt` reverted, but the manual edit in **agent-user.txt was also restored to HEAD**. This violates the file-list boundary and silently loses unrelated work. Please use `git --literal-pathspecs restore ...` (and the same protection for the `reset` branch), or literal-prefixed pathspecs, with tests covering `*`, `?`, brackets and pathspec-magic-looking filenames.

server/routes/git.js
```diff
@@ -1343,4 +1344,217 @@ router.post('/delete-untracked', async (req, res) => {
...
+  // per-file status are taken from the same instant.  Uses -z so NUL
+  // delimits records and rename pairs — this means a filename containing
+  // a literal " -> " cannot confuse the parser.
+  const statusByPath = new Map(); // path -> 2-char status (XY)
+  const renamedPaths = new Set();
+  let snapshotError = null;
+  try {
+    const { stdout: fullStatus } = await spawnAsync(
+      'git',
+      ['status', '--porcelain', '-z'],
```
davidliuk, line 1392: [P2] Enumerate untracked files instead of collapsed directories

Default `git status --porcelain -z` reports a new directory as `?? newdir/\0`, not its files. Reproduced by creating `newdir/agent.txt` and reverting exactly that file: the map contains only `newdir`, so the later lookup returns `success: true, reverted: [], skipped: ['newdir/agent.txt']` and leaves the agent-created file intact. This is common when an agent adds a module/test directory, and the UI labels skipped paths as unchanged. Add `--untracked-files=all` (keeping the separate directory-deletion refusal) and a regression for a requested file inside a completely new directory.

davidliuk (commented):
Ran the PR's existing targeted suites in an isolated checkout: **38 tests passed** (backend helper, HTTP route and file extractor). Temporary real-Git probes reproduced the two additional issues below without touching the working project.

There is also a scope mismatch with `Closes #85`: the issue explicitly requests the **exact state before the selected interaction**, not only preservation of unrelated files. I verified that HEAD=base, then a manual edit before the turn, then an agent edit, restores to base and loses the manual edit. The confirmation honestly warns about this and the PR already lists snapshots as a follow-up, so this is not an undisclosed implementation bug. However, please describe it as a Git-discard feature and use `Refs #85` rather than close the exact-undo issue, unless pre-turn snapshots and conflict detection are included. Later turns/manual edits to the same files also need protection for true per-turn undo.

PR #168, opened 2026-04-12
feat: rewrite managing-obsidian-notes skill v2.0 with CLI + onboarding

liuyixin-louis:
## Summary

- Rewrite `managing-obsidian-notes` skill from direct filesystem access to Obsidian CLI
- Full onboarding flow: app install → CLI enable → vault creation → PARA scaffold
- Per-project folder scaffolding with templates (project, paper, meeting)
- Always pops open notes in Obsidian after create/update
- No hardcoded personal paths — works with any user's vault
- Added to skills catalog (171 total) and all 5 pipeline stages

## Onboarding Flow

```
No Obsidian app → brew install --cask obsidian
No CLI         → Settings → General → Enable CLI → restart
No vault       → Create in Obsidian UI
Empty vault    → PARA scaffold: Projects/Areas/Resources/Archive/Inbox/Templates
New project    → Per-project folder: index, system-design, progress, key-links, experiment-log
```

## Vault Structure (PARA-inspired)

```
vault/
├── Home.md              ← Dashboard
├── Projects/            ← One folder per project
├── Areas/               ← Ongoing responsibilities
├── Resources/
│   ├── Papers/          ← Paper reading notes
│   └── Learning/        ← Tutorials, courses
├── Archive/             ← Completed work
├── Inbox/               ← Quick capture
├── Templates/           ← Project, Paper, Meeting templates
└── Attachments/         ← Media files
```

## Files Changed

| File | Change |
|------|--------|
| `skills/managing-obsidian-notes/SKILL.md` | Full rewrite (459 lines) |
| `skills/managing-obsidian-notes/CHANGELOG.md` | v2.0 entry |
| `skills/skills-catalog-v2.json` | Added entry (171 total) |
| `skills/stage-skill-map.json` | Added to all 5 stages |
| `skills/README.md` | Added Knowledge Management section to taxonomy |
| `.agents/skills/skills-index.md` | Added entry |
| `README.md` | Updated plugin section with MCP mention |

## Test plan

- [x] No PII in skill file
- [x] Skill registered in catalog and stage map
- [x] CLI commands match Obsidian CLI v1.8+ syntax

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #196, opened 2026-04-22
feat(skills): add writing-style skill for prose rule application

MarkZhang777:
## Summary

Adds `skills/writing-style/SKILL.md`, a drop-in skill that applies the 21 prose rules from the agent-style project (yzhao062, CC BY 4.0). The rule set is 12 canonical rules from Strunk, Orwell, and Pinker, plus 9 field-observed AI-output patterns.

Users invoke it with `/writing-style` in chat, optionally with a draft to revise. The skill surfaces in the Paper Writing category of the `⚡ Skills` picker alongside the complementary `/inno-humanizer`, which targets AI-specific tells rather than canonical prose rules. The existing `skillExpander.js` resolves the slash command, so no backend, UI, or database changes were needed.

Upstream: https://github.com/yzhao062/agent-style (pinned commit `6633397`).

## Test plan

- [ ] Inside an active chat, the `⚡ Skills` picker → Paper Writing category shows a `writing-style` chip alongside `inno-humanizer`, `inno-paper-writing`, etc.
- [ ] The expanded block renders the CC-BY-4.0 attribution to yzhao062/agent-style.
- [ ] `npm run typecheck` passes.


bbsngg (dismissed):
Small, self-contained, CI green. SkillDropdown renders straight from `SKILL_CATEGORIES` and the resolve route locates `SKILL.md` by directory, so `/writing-style` works in chat as-is. License and pinned upstream commit are recorded in the frontmatter. Approving.

One optional follow-up: the skill isn't registered in `skills/skills-catalog-v2.json` / `stage-skill-map.json` (compare #168), so it won't show in the Skills Dashboard. Running `scripts/export-skills-catalog-v2.mjs` should pick it up; fine to do in a follow-up PR.

bbsngg:
Pushed a follow-up commit on this branch registering `writing-style` in the catalog files (`skills-catalog-v2.json`, taxonomy overrides, `stage-skill-map.json`, `skill-tag-mapping.json`, and the `skills/README.md` tree), mirroring how `inno-humanizer` is registered, so the skill also appears in the Skills Dashboard and stage suggestions.

Heads-up unrelated to this PR: `scripts/export-skills-catalog-v2.mjs` currently fails on `main` because `skills/autoresearch/SKILL.md` has an unquoted `description:` containing `: `, which js-yaml rejects. I added the catalog entry by hand rather than regenerating (a full regen produces a ~9k-line diff against the committed catalog). Worth a separate one-line fix.

bbsngg (approved):
Re-approving after the catalog registration commit (86b6600a): CI green, all five JSON files parse.

PR #216, opened 2026-08-09
Feat/opus5 support

MarkZhang777:
## Summary

- Add `claude-opus-5` to the Claude model selector
- Configure its 1M-token context window for live and historical token usage
- Upgrade the Claude Agent SDK to 0.3.226

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -72,6 +72,39 @@ Platform mode is an advanced deployment option. Most users should leave these co
 
 ---
 
+## Model Discovery
+
+Dr. Claw asks each harness which models it actually supports rather than relying
+only on the list compiled into the app, so a CLI that ships a new model shows up
+without waiting for a Dr. Claw release.
+
+| Provider | Source | Notes |
+|----------|--------|-------|
+| Codex | `codex app-server` → `model/list` JSON-RPC | Same catalogue the Codex CLI's own picker reads. Honours `CODEX_CLI_PATH`. |
+| OpenRouter | `GET https://openrouter.ai/api/v1/models` | Public endpoint, no key needed. |
+| Claude, Cursor, Gemini, Nano | Built-in list | These CLIs expose no model-listing command today. |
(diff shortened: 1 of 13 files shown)


bbsngg:
Rebuilt this branch on top of `main` now that #213 is merged, so it no longer carries the discovery commit: `c3025f39` is the original Opus 5 change cherry-picked as-is (SDK 0.3.226, `claude-opus-5` entry, 1M context), and I pushed one commit on top.

**What the new commit does**

The Codex/OpenRouter discovery from #213 did not cover Claude, so the picker still depended on the hand-maintained table. The Agent SDK exposes the CLI's own menu via `query().supportedModels()`, so there is now a `claude` discoverer: a streaming-input session that never yields a prompt, reads the control-channel model list, and closes. No turn is consumed, `settingSources: []` keeps hooks from firing, and the session is closed on every path (verified no orphaned CLI process and no handle growth across repeated probes).

Measured with the SDK from this PR (0.3.226), the CLI serves: `default` (Opus 5, 1M), `opus[1m]`, `claude-fable-5[1m]`, `claude-fable-5-1[1m]`, `sonnet` (Sonnet 5), `haiku`. Latest (0.3.259) is the same minus `claude-fable-5[1m]`. Probe takes ~1-2.5s.

Because the CLI also runs ids it does not list (`claude-opus-4-6` etc.), Claude is flagged `acceptsUnlisted`: built-in entries stay undemoted, the configured default stands, and the picker's auto-rescue never fires for Claude. `[1m]` aliases now resolve to a 1M context window in both token-usage tables, and Fable 5.1 entries were added.

Tests: 31/31 in the discovery + route suites, typecheck clean, real end-to-end probe through the module works.

bbsngg (approved):
Approving: rebased onto main (single Opus 5 commit, SDK 0.3.226, 1M context) plus the Claude discoverer commit. CI green; 31/31 discovery + route tests and typecheck clean locally; real end-to-end probe verified.

PR #222, opened 2026-09-08
fix: complete loading review follow-ups and isolate lazy retries

davidliuk:
## Summary
Follow up on the remaining review items from #220:
- https://github.com/OpenLAIR/dr-claw/pull/220#pullrequestreview-5124295394
- https://github.com/OpenLAIR/dr-claw/pull/220#pullrequestreview-5125653772

### Workspace loading
- Remove the serial AppContent dynamic-import hop while keeping the survey lazy.
- Share the workspace loading fallback and preload workspace code from the login screen.
- Deduplicate pending/successful workspace imports, clear rejected promises, and catch speculative preload failures.
- Keep protected providers below ProtectedRoute so preloading code does not mount the authenticated workspace.

### Service worker
- Fetch each navigation once, cache only successful navigation/HTML responses, and extend cache-write lifetime with event.waitUntil.
- Return successful network responses without waiting for cache writes; cache failures do not fail the response.
- Retain the existing offline HTML fallback. This does not add full offline application support.

### Lazy loading recovery
Browser regression tests reproduced an infinite spinner caused by replacing React.lazy immediately when its import rejected, before the error could reach the boundary.
- Keep lazy identities stable through rejection, scoped to each LazyLoadBoundary.
- Rotate the scope only on explicit retry, an errored resetKey reset, or boundary remount.
- Preserve healthy component state and isolate sibling boundaries and late abandoned rejections.
- Provide an inner Suspense so the boundary-owned scope also commits with reversed outer-Suspense nesting.

Browsers may cache a failed module URL, so retrying the same URL is not guaranteed to refetch it. Persistent chunk failures expose Reload; recovery through Reload was verified.

### Regression coverage and CI
- Add six service-worker regression cases and three workspace-loader unit tests.
- Add six Chromium regressions for retry budget/recovery, resetKey and state preservation, sibling isolation, late rejection after remount, reversed nesting, and chunk-error UI.
- Wire Node regressions, Vitest, Chromium regressions, typecheck, and build into CI. Browser fixtures bundle the real source and need no backend or developer credentials.

## Verification
- Node Electron, service-worker, and browser regressions: **14 passed**.
- Vitest: **29 files, 202 tests passed**.
- Typecheck and git diff --check passed.
- Production build with manifest passed.
- Red/green reproduction: all six browser regressions failed before the lazy-retry fix and passed afterward.
- Production-browser checks with an isolated local mock: login preload, authentication, Dashboard, Trash, and intentional workspace-chunk failure followed by Reload recovery. No protected requests, WebSockets, or external requests before authentication; normal and recovered flows had no browser errors.
- Independent code review found no implementation blockers; stronger sibling-isolation assertions were incorporated and rerun.

Based on main at 8d36ccfe7a0bab7f6d1b61a04c73959234432ae0. Remote CI: build (20) and both CodeQL checks passed.


bbsngg (approved):
Reviewed the loading changes and the integration with current main. No blocking defects found in lazy retry isolation, reset/remount recovery, authenticated workspace preloading, or service-worker response/cache handling.

The LoginForm merge conflict is resolved while preserving main's initialMode/isFirstRun behavior and this PR's workspace preload.

Validation on 6a8b4cff: 30 Vitest files / 215 tests passed; 14 Electron, service-worker and Chromium regressions passed; TypeScript checking, production build and diff whitespace checks passed. Real-backend end-to-end flows and Firefox/Safari were not exercised in this review.

